PR #18, opened 2019-08-26
Solutions to Problem1&3 - Please Review

Arpan61:


Problem2.py
```diff
@@ -0,0 +1,40 @@
...
+
+        while queue and count < numCourses:
+            num = queue.pop(0)
+            for i in prerequisites:
+                if i[1] == num:
+                    arr[i[0]] -= 1
+                    if arr[i[0]] == 0:
+                        queue.append(i[0])
+                        count += 1
+        for i in arr:
```
sehaj-grover: why not use the count and just compare with numCourses

PR #32, opened 2019-09-18
Binary Tree level order and right side view program. LC verified

manishjangid:
Course Schedule in progress.  I had to implement dynamic queue to support the level order as there is no inbuilt library. 
It handles dynamic reallocation too.


kaurramanpreet:
I appreciate your efforts manish!!

PR #40, opened 2019-11-05
Not accepted on Leetcode. Work In Progress

sudhasr:


sudhasr:
Sure, hereafter I will make sure to update commits. 

Yes I am clear with these two programs now

sehaj-grover:
great!

rightSideView.py
```diff
@@ -0,0 +1,34 @@
...
+        :type root: TreeNode
+        :rtype: List[int]
+        """        
+        if root == None:
+            return []        
+        queue = []
+        result = []
+        queue.append(root)
+        while len(queue) > 0:
+            result.append(queue[-1].val)
```
sehaj-grover: what's the time complexity to get an element at index i from a list in python?

sudhasr: Time complexity to get an element at any index in a list is O(1) since we know the index, we can directly go there and access it.

sehaj-grover:
Clear with BFS-1?

sudhasr:
Hi Sehaj,

Now I understood that I have to use deque from collections rather than using a queue list for efficient time complexity. In queue list, pop(0)  time complexity would be O(N) because of the shifting of other elements. Whereas if I use deque, it would be O(1). I will implement using deque and submit the solution.

sehaj-grover:
It's correct

PR #54, opened 2019-11-13
Completed Binary Tree Level Order and Course Schedule

nshivapr:
Please Review

### DIFF
--- a/levelOrder.java
+++ b/levelOrder.java
@@ -1,0 +1,79 @@
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     int val;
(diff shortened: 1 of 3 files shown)


levelOrder.java
```diff
@@ -0,0 +1,79 @@
...
+            return;
+
+        if(result.size() == level)
+            result.add(new ArrayList<Integer>());
+
+        // get that temp list inside the result list for each level and add to that
+        result.get(level).add(root.val);
+        bfs(root.left, level+1);
+        bfs(root.right, level+1);
+    }
```
sehaj-grover: this is bfs or dfs

PR #57, opened 2019-12-23
Adding BFS changes

dasgprat:
Problem 2 is not solved completely. Seems some issue with my approach will do after discussion in class.

### DIFF
--- a/Problem1_Optimal.java
+++ b/Problem1_Optimal.java
@@ -1,0 +1,93 @@
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
+ *     int val;
+ *     TreeNode left;
+ *     TreeNode right;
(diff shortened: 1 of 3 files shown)


dasgprat:
Added problem 2 after in-class discussion

Problem3_DFS.java
```diff
@@ -0,0 +1,43 @@
...
+Time Complexity : O(N)
+
+The idea is to recursively traverse right side of the tree and check if the depth of the tree == size of the output list - 1. if depth matches size of the output list - 1 add the node to the output list.
+**/
+
+class Solution {
+    List<Integer> output = new ArrayList<>();
+    public List<Integer> rightSideView(TreeNode root) {
+        
+        if(root == null)
```
neha79091: Good work prateek

PR #61, opened 2019-12-29
Completed BFS 1

RishikaMachina:


Problem_2.py
```diff
@@ -0,0 +1,60 @@
+# Runs on Leetcode
+# Runtime - O(E) - E is edges
+# Memory - O(n)
+
+'''
```
neha79091: here complexity will be O(V+E)

Problem_3.py
```diff
@@ -0,0 +1,48 @@
+# Runs on Leetcode
+# BFS
+    # Runtime - O(n)
+    # Memory - O(n)
+# DFS
+    # Runtime - O(n)
+    # Memory - O(1)
+
```
neha79091: Space complexity here will be O(h), if you include the stack

RishikaMachina: Stack as in recursion stack?

Problem_2.py
RishikaMachina: Changed!

Problem_3.py
neha79091: yup

RishikaMachina: Should we consider it?

neha79091: you can consider it by saying that you are considering the recursion stack

PR #63, opened 2019-12-29
Added solutions for CourseSchedule, RightSideView and LevelOrderTrave…

amitkc1:
…rsal

### DIFF
--- a/LevelOrderTraversal.java
+++ b/LevelOrderTraversal.java
@@ -1,0 +1,43 @@
+//Time Complexity: O(n)
+//Space Complexity: O(n)
+//Solved on Leetcode: Yes
+
+/**
+ * Definition for a binary tree node.
+ * public class TreeNode {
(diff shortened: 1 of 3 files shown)


LevelOrderTraversal.java
```diff
@@ -0,0 +1,43 @@
+//Time Complexity: O(n)
+//Space Complexity: O(n)
+//Solved on Leetcode: Yes
+
```
neha79091: Space complexity will be O(h)

RightSideView.java
```diff
@@ -0,0 +1,40 @@
+// Time Complexity: O(n)
+// Space Complexity: O(1)
+// Solved on leetcode : Yes
```
neha79091: Space complexity here will not be constant

neha79091: it will be O(h)

PR #72, opened 2020-01-22
Please review my Solutions

Sanjay-995:


BigN58RightSideViewUsingBFS.java
```diff
@@ -0,0 +1,33 @@
+// Time complexity is O(N) as each node is being traversed
+// space complexity is O(1) as no extra space is used.
+// This solution is submitted on leetcode
```
neha79091: We are using queue here ,so space complexity will not be constant.

BigN58RightSideViewUsingBFS.java
```diff
@@ -1,5 +1,5 @@
 // Time complexity is O(N) as each node is being traversed
-// space complexity is O(1) as no extra space is used.
+// space complexity is O(N) as we are using queue data structure
```
neha79091: Good

PR #107, opened 2020-04-13
Completed first question of BFS-1

KBangera:
Will complete and submit the other questions soon


devashishnyati:
Good job Karthik!
Submit the other problem as well.

PR #115, opened 2020-04-14
Completed BFS 1 Please Review

kannan-mrugapphan:
passes all test cases on leetcode

### DIFF
--- a/courseSchedule.java
+++ b/courseSchedule.java
@@ -1,0 +1,67 @@
+// 207.
+
+//time - O(n)
+//space - O(n) for the indegree array, hashmap and queue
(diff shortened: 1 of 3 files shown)


courseSchedule.java
```diff
@@ -0,0 +1,67 @@
+// 207.
+
+//time - O(n)
```
devashishnyati, line 3: Time Complexity:O(2⋅∣E∣+∣V∣) where ∣V∣ is the number of courses, and ∣E∣ is the number of dependencies. -Leetcode

devashishnyati:
Good job Kannan!

PR #121, opened 2020-04-18
BFS 1 problems solved

DebanjanaDasgupta:


Problem3.java
```diff
@@ -0,0 +1,55 @@
...
+ *     int val;
+ *     TreeNode left;
+ *     TreeNode right;
+ *     TreeNode(int x) { val = x; }
+ * }
+ */
+
+
+// Recursive solution (no optimization)
+// Time complexity: O(n+m), n = number of nodes. m = no of edges
```
devashishnyati, line 13: Why is it O(m+n)? You are visiting each node only once.

devashishnyati:
Good job Debanjana!

PR #139, opened 2020-05-21
PLease Review

sbsreedh:


levelOrder.py
```diff
@@ -0,0 +1,81 @@
...
+# class TreeNode:
+#     def __init__(self, val=0, left=None, right=None):
+#         self.val = val
+#         self.left = left
+#         self.right = right
+#BFS Solution:
+#Time Complexity:O(n)
+#Space Complexity:O(n)
+#Ran successfully on Leetcode:Yes
+#Difficulties: Since I am using nested for loop while returning result, does this make my TC as O(N^2)??
```
iivaishnavii, line 11: We do not find time complexity through loops. In case of trees, we look at how we cover the tree nodes

iivaishnavii:
Good job with all the 3 problems.. Keep up the consistency!!

PR #141, opened 2020-05-21
Create sample.java. Please review BFS-1

Abinashi7:


iivaishnavii:
Great job with all the three problems, Abinashi! Keep it up.. Don't forget to try out righside with dfs as well!

PR #144, opened 2020-05-22
PA55, P56, and PA59 solution

HimanshuAgarwal93:
PA55:
//Leetcode 102. Binary Tree Level Order Traversal
//BFS solution
//Time Complexity : O(n)  // n is the number of nodes in the tree
//Space Complexity: O(n) //size of Queue, where size is equal to max no nodes at a level in tree i.e n/2

PA56:
//Leetcode: 207. Course Schedule
//Time Complexity: O(n) // where n is no of courses
//Space Complexity: O(n) // n+n n for hashmap and n for queue

PA59:
//Leetcode 199. Binary Tree Right Side View
//Time Complexity : O(n)  // n is the number of nodes in the tree
//Space Complexity: O(n) // size of Stack, where size is equal to height of the tree i.e log(n)

### DIFF
--- a/PA55.java
+++ b/PA55.java
@@ -1,0 +1,54 @@
+//Leetcode 102. Binary Tree Level Order Traversal
+//BFS solution
+//Time Complexity : O(n)  // n is the number of nodes in the tree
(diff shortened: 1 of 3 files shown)


iivaishnavii:
Good job, Himanshu!!

PR #156, opened 2020-06-23
BFS-1 . Please review

Pradhisha:


BT_RightSideView.java
```diff
@@ -0,0 +1,27 @@
+// Time Complexity :O(n) n - number of tree nodes
+// Space Complexity :O(2^h) h - height of the tree
```
Lohith-Aswathappa, line 2: Since you are using the BFS approach here, Height of the tree doesn't come into picture.

Lohith-Aswathappa: check the space complexities again

binaryLevelOrderTraversal.java
```diff
@@ -0,0 +1,70 @@
+// Time Complexity :O(n) n - number of tree nodes
+// Space Complexity :O(2^h) h - height of the tree
```
Lohith-Aswathappa, line 2: check space complexity here again

Lohith-Aswathappa:
good work, Pradhisha!
Review the inline comment.

BT_RightSideView.java
Pradhisha: I am using queue which has nodes of every level. space complexity of queue 2^h. This was idea behind the Space complexity

PR #169, opened 2020-06-24
BFS-1 solved Please review

umamnataraj9:
BFS-1 solved Please review


AbhiniveshP:
For binary tree questions, try implementing the DFS for both of them.
For course schedule, please explicitly mention what n is?
There are two things to consider here, number of courses and number of prerequisite pairs.

AbhiniveshP:
For binary tree questions, try implementing the DFS for both of them.
For course schedule, please explicitly mention what n is?
There are two things to consider here, number of courses and number of prerequisite pairs.

PR #170, opened 2020-06-24
Completed-Please Check

DineshNadar95:


AbhiniveshP:
For level order question, time complexity is O(n) ad not O(h) if a queue is used.
For course schedule, when time complexity is mentioned in terms of n, what is n here? Is it the number of courses or the prerequisite pairs?

AbhiniveshP:
For level order question, time complexity is O(n) ad not O(h) if a queue is used.
For course schedule, when time complexity is mentioned in terms of n, what is n here? Is it the number of courses or the prerequisite pairs?

PR #177, opened 2020-06-24
Please Review

soumitra9:


right_view_BFS.py
```diff
@@ -0,0 +1,35 @@
+# Time Complexity : Add - O(n)
+# Space Complexity :O(2^h)
```
Lohith-Aswathappa, line 2: in BFS you'll be traversing according to level, explain the what the variable H is here

Lohith-Aswathappa:
well done, Soumitra!
Review the inline comment.

right_view_BFS.py
soumitra9: Yeah, the maximum number of nodes in the queue would be the width(or leaf nodes) which is proportional to 2^h. By h I meant to imply the number of levels.

PR #194, opened 2020-06-30
BFS-1-2Problems--Completed

VirajDeshmukh1:


AbhiniveshP:
Nice work!!!
Can you implement Course Schedule using both DFS and BFS? What's the difference between detecting a cycle in a directed graph, topological sort, and the current course schedule problems? Can you answer these questions?
Also, try implementing the BFS version (both left first, right first) for right side view question.

AbhiniveshP:
Nice work!!!
Can you implement Course Schedule using both DFS and BFS? What's the difference between detecting a cycle in a directed graph, topological sort, and the current course schedule problems? Can you answer these questions?
Also, try implementing the BFS version (both left first, right first) for right side view question.

PR #222, opened 2020-08-10
BFS-1

SantoshReddyKallala:


CourseSchedule.java
```diff
@@ -0,0 +1,45 @@
...
+                    }
+                }
+            }
+        }
+
+        if(count==inedge.length) return true;
+
+        return false;
+    }
+    //Time COmplexity : o(c) c->number of courses
```
samarthjj, line 42: The time here will also be O(V + E)

samarthjj:
Good work Santosh, Please review the inline comments. 
Please also try solving **Level Order Traversal** & **Right-Side View problems** with the DFS solution as well.

samarthjj:
Good work Santosh, Please review the inline comments. 
Please also try solving **Level Order Traversal** & **Right-Side View** problems with the DFS solution as well.

samarthjj:
Good work Santosh, Please review the inline comments. 
Please also try solving **Level Order Traversal** & **Right-Side View** problems with the DFS solution as well.

PR #265, opened 2020-10-04
BFS-1 Complete-Needs Review

nmtraboulsi:
First, pull request to S30.


BInaryTreeRightSideView.java
```diff
@@ -0,0 +1,55 @@
+import java.util.*;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+/**
+* Time Complexity: O(N), it has to visit each node in the tree
+* Space Complexity: O(D), where D is the tree's diameter
```
samarthjj, line 6: The asymptotic space complexity here would be O(n), where n is the total number of nodes in the tree.
In a typical BFS traversal, the maximum amount of nodes/objects the queue holds is all of the last level of the tree. 
For a balanced tree the last row precisely has n/2 nodes, which also happens to be the worst case.

samarthjj:
Good work here Nicole, Please review the inline comments. Good job on completing your first Pull Request!
Going further please also add a 3-4 sentence explanation of your approach for each problem, this is essentially how you would explain your solution in an interview.

samarthjj:
Good work here Nicole, Please review the inline comments. Good job on completing your first Pull Request!
Going further please also add a 3-4 sentence explanation of your approach for each problem, this is essentially how you would explain your solution in an interview.

PR #285, opened 2020-10-14
bfs-1

rajjiyani:


116. Populating Next Right Pointers in Each Node.java
```diff
@@ -0,0 +1,30 @@
+class Solution {//Time and space of O(n)
+    // revise look other approach
```
kannan-mrugapphan, line 2: Please try the recursive approach as well

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #410, opened 2021-02-16
BFS-1 done. Please review!

EnnaSachdeva:


course_schedule.py
```diff
@@ -0,0 +1,51 @@
+
```
nsibal, line 1: Please mention Time and Space Complexities and your approach in brief.

nsibal (commented):
Looks good!
Read inline comments.

nsibal:
Looks good!
Read inline comments.

PR #427, opened 2021-03-17
Completed BFS-1 Please review

Mani632:


kannan-mrugapphan (commented):
Good Job. Please recheck on the time for the 1st problem. Its O(v + e). The space for the second problem is O(n) for the queue. Use a queue and not a list for bfs

kannan-mrugapphan:
Good Job. Please recheck on the time for the 1st problem. Its O(v + e). The space for the second problem is O(n) for the queue. Use a queue and not a list for bfs

PR #430, opened 2021-03-17
Completed Problem 1

Tirthaj:


Lohith-Aswathappa (commented):
Good work, Tirthaj!
try the other problem as well.

Lohith-Aswathappa:
Good work, Tirthaj!
try the other problem as well.

PR #703, opened 2022-02-27
BFS-1 Completed

priyanshi9692:
Ready for review

### DIFF
--- a/CourseSchedule.java
+++ b/CourseSchedule.java
@@ -1,0 +1,54 @@
+import java.util.*;
+
+/*
+Time Complexity:O(v+e)
+Space Complexity: O(n)
+Run on leetcode: yes
+Any difficulties: no
+
+Approach:
+1. Finding out a cycle in the graph, is cycle is there, there is no chance to complete the courses
+2. It should be directed acyclic graph
+ */
+public class CourseSchedule {
+    public static boolean courseSchedule(int[][] prerequisites, int numCourses ){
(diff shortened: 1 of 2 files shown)


CourseSchedule.java
```diff
@@ -0,0 +1,54 @@
+import java.util.*;
+
+/*
+Time Complexity:O(v+e)
+Space Complexity: O(n)
```
shrivatsav, line 5: Please specify what `v,e,n` indicate.

LevelOrderTraversal.java
```diff
@@ -0,0 +1,65 @@
...
+/*
+Time complexity: O(H)- height of the tree
+Space Complexity: O(N)
+Run on leetcode: yes
+Any difficulties: No
+
+Approach:
+1. Recursively add values at root and increase the level
+ */
+public class LevelOrderTraversal {
```
shrivatsav, line 13: Please describe your approach briefly.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

CourseSchedule.java
priyanshi9692: N in the space complexity indicates use of Queue, and Adjacency List length,
v denotes number of vertices of the graph which is number of courses in this case
e represents number of edges of the graph which in this case indegree of every vertex

PR #818, opened 2022-06-22
BFS-1 Completed - Please Review

Ruthwik12:


Binary Tree Level Order Traversal.py
```diff
@@ -0,0 +1,52 @@
+# time complexity is o(n), where n is the size of the input
+# space complexity is o(n), where n is the size of the input
```
daredevilteja, line 2: SC should be O(n/2) because the maximum size of the queue will be leaf nodes

daredevilteja (commented):
Good Work! Keep it up

daredevilteja:
Good Work! Keep it up

PR #896, opened 2022-08-20
bfs...please eview

Sreelekhasamanu:


course_schedule.java
```diff
@@ -0,0 +1,47 @@
+//TC: O(n)
```
daredevilteja, line 1: TC is O(V+E)

course_schedule.java
```diff
@@ -0,0 +1,47 @@
...
+//SC: O(n)
```
daredevilteja, line 2: SC is O(V+E)

daredevilteja:
Good Work! Keep it up

PR #966, opened 2022-11-06
BFS 1, Both recursive and BFS approaches.

chaithanya2sai:


CourseSchedule.java
```diff
@@ -0,0 +1,55 @@
+package s30.BFS;
+
+import java.util.*;
+
+
+// TIme Complexity O(m) m being the number of rows of prerequisites
```
daredevilteja, line 6: PLease recheck the TC and SC

daredevilteja:
Good Work! Keep it up.

PR #1188, opened 2023-12-31
Completed BFS-1

kiruthikaa:


LevelOrderTraversal.java
```diff
@@ -0,0 +1,47 @@
+//Time complexity:O(n)
+//Space Complexity:O(h), height of binary tree
```
ishika2741, line 2: It would be O(n) for BFS because width of tree goes inside the queue

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1225, opened 2024-02-21
Done with BFS-1

Prinitha:


BinaryTreeLevelOrderTraversal.py
```diff
@@ -0,0 +1,27 @@
+'''
+TC: O(n) where n is number of nodes
+SC: O(1)
```
ishika2741: Check space

ishika2741 (commented):
Good work!

ishika2741:
Good work!

BinaryTreeLevelOrderTraversal.py
Prinitha: Right! Just corrected. Since it is DFS, the space should be O(h) - h being the height of the tree.
Thanks.

PR #1240, opened 2024-03-02
BFS1

nguyenmv2:


Nandha2Tech (commented):
Great Work! SC for level order traversal in BFS would be O(n/2) -> O(n).

Nandha2Tech:
Great Work! SC for level order traversal in BFS would be O(n/2) -> O(n).

PR #1305, opened 2024-07-29
completed BFS-1

dhikki-study:
#######102. Binary Tree Level Order Traversal###################################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
I have used here BFS approach where we use a queue to store element on same level and process them on level basis

        
        

######207. Course Schedule################################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : NA

// Your code here along with comments explaining your approach in three sentences only
Here we have also taken BFS approach where we stored all the courses in queue with the prerequisites also we have hash map to map the course to its dependent. So when ever any courses reaches requirement of zero it is released for queue and we keep doing it till our count matches courses

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,104 @@
+#######102. Binary Tree Level Order Traversal###############################################################################################################
+// Time Complexity : n
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work! TC and SC for course schedule would be O(V+E).

Nandha2Tech:
Great Work! TC and SC for course schedule would be O(V+E).

PR #1340, opened 2024-10-22
Solved Problem-1 and 2

nitishthorat:


Nandha2Tech (commented):
Great Work! Time and space for can finish the courses problem would be O(V+E).

Nandha2Tech:
Great Work! Time and space for can finish the courses problem would be O(V+E).

PR #1358, opened 2024-11-24
COmplete BFS-1

Thaslim:


Erlybird:
For the level order traversal, if the input is an empty tree, you should return a list with 0 elements, but you are returning null.

Erlybird:
For the level order traversal, if the input is an empty tree, you should return a list with 0 elements, but you are returning null.

PR #1546, opened 2025-07-21
BFS-1

thakuryash060:


super30admin:
- Problem1.java: Your DFS approach for level order traversal is correct and demonstrates good understanding of tree traversal. However, note that BFS is more conventional for this problem as it naturally processes nodes level by level.
- Problem2.java: You have a critical error in the queue initialization where you add `indegree[i]` instead of `i`. This will cause the algorithm to fail. Correct this by adding `i` to the queue when `indegree[i] == 0`. Additionally, ensure you handle cases where `map.get(curr)` might return null to avoid NullPointerException.

Overall, your understanding of the algorithms is good, but attention to detail in implementation is crucial. Always test edge cases to catch such mistakes.

PR #1592, opened 2025-11-01
BFS-1 complete

PrasiddhShah:


super30admin:
- Problem1.java: Well done! Your BFS implementation is correct and efficient. The comments are clear and helpful.
- Problem2.java: Your approach is on the right track but needs improvement. The main issue is that you're not correctly checking if x and y have the same parent. You should track both depth and parent information to accurately determine if nodes are cousins. Also, the space complexity should be O(h) due to recursion stack, not O(1). Consider modifying your DFS to track parent nodes as well.

PR #1600, opened 2025-12-10
Completed BFS-1

pranavs0133:


super30admin:
Strengths:
- The student has implemented both solutions correctly and efficiently.
- Appropriate data structures are used for both problems.

Areas for Improvement:
- In Problem1, the commented-out DFS method is unnecessary and could be removed to keep the code clean.
- In Problem2, the initial check `if(cnt == numCourses)` is redundant because if all courses have indegree 0 initially, the while loop will not execute, and the final check will handle it. This could be simplified.
- The import statement `import javax.swing.tree.TreeNode;` in Problem1 is incorrect. It should be the custom TreeNode class or the correct import from the problem definition.

PR #1618, opened 2026-01-25
BFS-1 completed

sujaygijre:


super30admin:
Your solution is well-written and efficient. You have correctly implemented the level order traversal using a queue. The code is clear and concise. One minor suggestion: you could use a for loop instead of a while loop with decrement for the inner loop to make it more explicit, but the current way is also acceptable. Great job!

PR #1623, opened 2026-02-03
1 problem completed

BharathVuppala96:


super30admin:
Your solution is correct and follows the BFS approach appropriately. You have correctly identified the level boundaries by using the size of the queue at the beginning of each level. The code is clean and easy to understand.

One area for improvement is the efficiency of the queue operations. In Python, lists are not efficient for popping from the front because each pop(0) requires shifting all subsequent elements, which takes O(n) time. For large trees, this could lead to performance issues. Instead, you should use `collections.deque` which is designed for efficient appends and pops from both ends. With a deque, popping from the left is O(1), making the overall algorithm O(n).

Here's how you can modify your code:
1. Import deque: `from collections import deque`
2. Initialize the queue as a deque: `queue = deque([root])`
3. Instead of `queue.pop(0)`, use `queue.popleft()`

This change will optimize the time complexity to O(n) as intended.

Other than that, your solution is good. Keep up the good work!

PR #1635, opened 2026-03-14
Level order, course schedule

shaikhmisrail:


super30admin:
### Level Order Traversal in Binary tree (leetcode-102.py)
Your solution is well-structured and follows the standard approach for level order traversal. Here are some points of feedback:

Strengths:
- You correctly use a queue (deque) for BFS, which is efficient for this problem.
- You handle the base case appropriately by returning an empty list if the root is null.
- You correctly use a for-loop to process each level by capturing the queue size at the start of the level.
- The code is readable and follows good naming conventions.

Areas for Improvement:
1. Import Statement: You import `deque` from `collections`, which is correct. However, it's a good practice to be explicit about imports. Alternatively, you could use `from collections import deque` as you did, but ensure consistency. Your code already does this correctly, so no issue here.
2. Initialization: You initialize `q = deque([])`, which is equivalent to `q = deque()`. The latter is slightly more concise and equally clear.
3. Efficiency: The solution is efficient, but note that using `deque` is efficient for popping from the left (O(1)), which is important for BFS. This is better than using a list which would have O(n) for popping from the left.
4. Code Comments: While the code is clear, adding comments to explain each step (especially for complex problems) can be helpful. However, for this straightforward solution, it might not be necessary.

Overall, your solution is correct, efficient, and follows good practices. Well done!

VERDICT: PASS

---

### Scheduling Courses (leetcode-207.py)
Your solution is correct and efficient. You have implemented Kahn's algorithm correctly. Here are some suggestions for improvement:
- Update the comments to reflect the correct time and space complexity: O(V+E) for both, not O(n). Since V (numCourses) and E (number of edges) are the key variables.
- Consider renaming `el` to something more descriptive like `prereq` or `edge` to improve readability.
- The initial check `if len(q) == numCourses:` is not necessary. You can remove it without affecting correctness. The main loop and the final check for non-zero in-degrees are sufficient.
- It's good practice to add a docstring to the method explaining what it does.

Overall, your code is clean and efficient. Well done!

VERDICT: PASS

